Issue #12, opened 2022-05-02
Decouple Dalamud and XL in terms of branding 


philpax:
There's several places where Dalamud is referred to as "XL plugins" or "in-game features", which leads to user confusion. We should consider making the distinction more obvious and deprecating the old commands/terminology.

PR #17, opened 2022-05-03
Final Comment: Automated build and submit pipeline

philpax:
First DIP, yay!

This is a first draft with a few unanswered questions. I've left them vague so that we can specify them.

Would like input from @ayyaruq and @NotAdam if possible.


text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+
+The primary role of the action is to consume a (repository, commit) and produce an artifact that can be loaded by Dalamud, or be included within DalamudPlugins. DalamudPlugins will only be updated on deployment (e.g. committing of the metadata file to `main` in the repository); it will not be updated otherwise.
+
+(RFC: I feel like I'm forgetting something here...)
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+This will require all first-party plugin developers to abide by a common project structure that is amenable to being built by CI, and for the code of their plugins to be open-source. The latter is already true, but the former may be a difficult requirement, especially with some of the complex build arrangements that existing plugins use.
```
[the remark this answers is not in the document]
NotAdam: probably should have some way of surfacing what's used easily so someone can look and see if there's anything amiss quickly more than anything else

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+This design has a few key benefits over competing designs:
+
+- We don't pay for compute
+- It is fully declarative (e.g. what's specified in the metadata repo is what's deployed)
+- It does not require us to trust plugin developers' build systems
+- It is not a significant change from our existing system
+
+Initial discussion proposed a web portal, likely in addition to existing goatcorp services, where developers could submit their plugins to be reviewed. While this would be nice to have, it's not really immediately necessary.
+
+An alternative design was proposed by [Ay](https://github.com/ayyaruq) where plugin developers, upon main repo approval, would be given keys exclusive to the plugin that could be used to sign future updates. This would allow new versions of plugins to be deployed without explicit goatcorp approval. However, this solution was put aside due to its relative complexity for limited payoff.
```
NotAdam, line 84: i think this is too complicated and not worth it

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+
+RFC: Got some cool prior art? Ay mentioned working on something similar in the past.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- What should the directory/project structure look like for a plugin under this system? Should we even be enforcing it?
+- What should be allowed in a plugin `csproj`?
+- Should the developer specify the packaging step within their `csproj`, or should we do it for them?
```
NotAdam, line 100: plugin devs should define their plugin metadata in the csproj (basically how dalamudpackager works) and we use the same props

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+- Do we want to automatically approve updates to already-submitted plugins?
```
NotAdam, line 101: depends how much you trust people, but then you can get 1 good release in and then immediately release malware, so lol

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+This would form the backbone of any future CI we do around plugin submission. Ideas for improvement include:
+
+- Implementing a web frontend for the plugin submission/review/update process. Instead of manually PRing up changes to the metadata repo, a plugin developer could pop over to the frontend and specify which commit they'd like deployed, then hit the button.
+- As we're building the code, we have an opportunity to analyse it and ensure that it isn't doing anything outright naughty. We would likely use a solution similar to [S&box](https://wiki.facepunch.com/sbox/AccessList)'s. This could be combined with other steps, like asking the plugin to declare the namespaces it will use.
```
NotAdam, line 110: this doesn't account for unsafe code at all, which basically lets you do whatever you want

NotAdam (changes requested):

text/0000-automated-build-and-submit-pipeline.md
kalilistic, line 101: I think an idea before was auto-approving testing plugins but not main.

text/0000-automated-build-and-submit-pipeline.md
kalilistic, line 69: Could review existing plugins to see how many would have significant issues under this.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+The `repository` refers to a Git repository (not necessarily hosted on Git), and the `commit` refers to a specific commit to build and deploy. For this to work, the repository must be structured to be amenable to automatic builds and packaging. Details of this will be covered in [reference-level-explanation].
+
+When submitting a plugin for the first time, the developer creates the file with the relevant details filled in, and creates a PR. A GitHub Action will retrieve the repository at the specified commit, attempt to build it, and produce an artifact that can be downloaded and loaded into Dalamud for testing by a goatcorp plugin reviewer. If accepted, the PR is merged, and the deployment process follows.
+
+When a new version of a metadata file is merged, a GitHub Action will build the repository, as with initial submission, and the resulting artifact will be deployed to DalamudPlugins automatically. The developer does not need to do anything after the metadata update has been merged.
+
+When updating a plugin, the developer merely needs to update the `commit` for the metadata file, and make a PR. The previously mentioned steps will take place.
```
hedgeye-fslezak, line 45: How would this handle changelogs for plugins? For example, I may want to include testing notes or create a "rollup" of my changes, or otherwise diverge from "these are the git commits between these times" as these are often technical and not necessarily good/useful to end users.

KazWolfe, line 45: How would this handle changelogs for plugins? For example, I may want to include testing notes or create a "rollup" of my changes, or otherwise diverge from "these are the git commits between these times" as these are often technical and not necessarily good/useful to end users.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+A new GitHub repository will be created. In this repository, each file corresponds to a submitted plugin and contains a JSON/TOML/YAML structure that looks something like this:
+
+```json
+{
+    "repository": "https://github.com/philpax/plogonscript.git",
+    "commit": "7ba9de4049066218f67e260e2c0f8ec0651ffed7"
+}
+```
+
+The `repository` refers to a Git repository (not necessarily hosted on Git), and the `commit` refers to a specific commit to build and deploy. For this to work, the repository must be structured to be amenable to automatic builds and packaging. Details of this will be covered in [reference-level-explanation].
```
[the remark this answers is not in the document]
NotNite: "not necessarily hosted on Git" doesn't make much sense - did you mean GitHub here?

NotNite:
Would a developer be able to test the build process to make sure it works before PRing? One concern of mine is merging a PR and then the plugin build fails in the action. Perhaps we could provide a standalone action runnable through [act](https://github.com/nektos/act), or a Docker container.

~~also can we please make a cool dalamud logo ascii art for the script that'll build the plugins~~

NotNite:
Also, it should be of note we set up dependencies properly. The script should make sure to clone Git submodules and install NuGet packages. We should also have it post webhooks to `#github` so we'll know if a build is failing.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,111 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+A new GitHub repository will be created. In this repository, each file corresponds to a submitted plugin and contains a JSON/TOML/YAML structure that looks something like this:
+
+```json
+{
+    "repository": "https://github.com/philpax/plogonscript.git",
+    "commit": "7ba9de4049066218f67e260e2c0f8ec0651ffed7"
```
NotNite, line 35: What about testing? Will it have a separate folder, or a `testing_commit` entry in this file?

karashiiro:
That shouldn't be an issue, this wouldn't run any plugin code. If the deployed code is a build requirement, that could be done as a build task, but if it's a runtime requirement then it shouldn't matter.

karashiiro:
Of course, that also partially-undermines the point of this DIP, which is something else to think about.

KazWolfe:
> if it's a runtime requirement then it shouldn't matter.

I'd tend to agree, but I do want to bring it up because it does add a bit of thought to how/when/why things would be triggered on the Dalamud side for devs of plugins like this. These plugins are more likely to do Other Things™ as well that may cause varying degrees of headaches under the current proposal:

* Both Ava and I place all of our components in our respective repositories. We'd need to have some way to tell the build system which `.csproj` file to use and which to ignore. While a forced repository restructure would work here, I'm hesitant to propose that as a solution.
* It is completely possible for dependencies/other components to require different build environments for one reason or another. The Stream Deck Plugin, for example, builds inside of a Ubuntu GitHub Action. A client/server model may want to build with different dependencies in a different context. I don't think it's reasonable to push these build deps over to Goatcorp, nor require they handle external CI.
* As alluded to, plugins that inherently depend on external things (be they build-time or runtime) already somewhat break the entire point of this DIP. Goatcorp doesn't (can't) know what the library I'm building is doing necessarily, especially if the build is grabbing it from my Releases page in real-time. Similarly, a client-server plugin with a runtime dep can have behavior changed at any time.

If the solution to this problem becomes "trigger your own CI, do whatever you need to do, and then automatically submit a PR once you're ready", that's perfectly acceptable (at least to me). 

> perhaps they could be moved to the "blessed second-party repositories" we've been talking about in a separate DIP.

If this becomes a thing, I'd like some way for my plugin(s) to still show to all users for discovery purposes. Users "window shop" for plugins a lot; I'd hazard a guess that a fair number of installs come from "oh, cool, there's a plugin that can do this?" rather than specifically seeking out a function. Blessed repos should be default-visible, but I suppose that's a discussion for that DIP instead of this one

text/0000-automated-build-and-submit-pipeline.md
KazWolfe: Mm, I suppose that could work but I'm not too big of a fan on that (though this comes from a "my own use case" model). Generally, I dislike the idea of changelogs being files in the repository as now that creates three different "views" of history:

* The `CHANGELOG.md` (or whatever) file that lives in the repo proper and will be displayed in Dalamud
* The commit messages themselves, as well as tag annotations (if any)
* The information available in GitHub Releases itself

Broadly speaking these should be *mostly* in sync, but some formatting/stylistic differences will appear. Text that renders well in GitHub Releases may not render as nicely in-game or in Discord. Information in one place may not necessarily apply to others.

Granted, this is already a problem, but it can be resolved through manual control of the release notes (e.g. just creating different copy-edits for the PR and the GH Releases page).

philpax: What would be your ideal solution? Something that comes to mind is taking the description of the PR, or a subset of that (e.g. the PR contains a special "changelog" section).

philpax:
> Would a developer be able to test the build process to make sure it works before PRing? One concern of mine is merging a PR and then the plugin build fails in the action. Perhaps we could provide a standalone action runnable through [act](https://github.com/nektos/act), or a Docker container.
> 
> ~also can we please make a cool dalamud logo ascii art for the script that'll build the plugins~

Fine by me! I'm also entirely fine letting people fail their builds on submission, failures happen all the time. Just need to make sure they're graceful :)

I'll make a note about making sure the action is compatible for local running...

text/0000-automated-build-and-submit-pipeline.md
philpax: Good catch. Separate folder, similar to how `DalamudPlugins` currently works:
```
main/plogonscript.json
testing/plogonscript.json
```

text/0000-automated-build-and-submit-pipeline.md
philpax: Decided to just add `changelog` to the submitted manifest. Seems simplest, and is nice and declarative.

philpax:
I've updated the DIP to address all of the feedback mentioned here, as well as discussion in Discord. I expect the proposed changes to the `csproj` to be a little controversial, but the rest should be more or less acceptable. Let me know what y'all think!

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -27,51 +27,121 @@ Upon completion, this should allow for developers to easily submit new plugins f
...
+
+- `main`: The primary repository for Dalamud plugins, and what first-party plugin users will have access to. This is the most stringest to submit to, as it has the widest audience, but we should offer all the tools we can to reduce the barrier to entry.
+- `testing`: A repository for plugins that still require additional testing, or don't meet first-party requirements yet. The requirements to submit to `testing` are lower than that of `main`, but plugins still have to follow the general guidelines (e.g. no bots!)
+  - Additionally, to assist in the testing process, it is likely that update (not submission) PRs to `testing` will automatically be approved and merged. This allows developers to rapidly test changes, but ensures that human oversight will still be in place for when they submit to `main`.
+- `net6`: Plugins compiled for the .NET 6 version of Dalamud, so that they can be tested for the migration.
```
NotAdam, line 35: `testing/net6` instead? `net6` by itself doesn't really mean anything by itself in terms of where it lives in the release pipeline

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -27,51 +27,121 @@ Upon completion, this should allow for developers to easily submit new plugins f
...
+            <Tag>Gathering</Tag>
+            <Tag>Fishing</Tag>
+            <Tag>Miner</Tag>
+            <Tag>Botanist</Tag>
+            <Tag>Weather</Tag>
+            <Tag>Alarms</Tag>
+            <Tag>Timer</Tag>
+        </Tags>
+        <Hidden>False</Hidden>
+        <TestingExclusive>True</TestingExclusive>
```
NotAdam, line 78: what's the point of this if it only exists on the testing branch? isn't this a process failure if it ends up on main?

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -95,10 +210,14 @@ RFC: Got some cool prior art? Ay mentioned working on something similar in the p
...
 
-- What should the directory/project structure look like for a plugin under this system? Should we even be enforcing it?
-- What should be allowed in a plugin `csproj`?
-- Should the developer specify the packaging step within their `csproj`, or should we do it for them?
-- Do we want to automatically approve updates to already-submitted plugins?
+Several points within this DIP were discussed during Discord conversations. Here are links, as well as a brief summary:
+
+- <https://discord.com/channels/581875019861328007/860813266468732938/972255490128113684>
+  - Concerns raised during this discussion:
+    - **Not every dev wants to use DalamudPackager**: they won't need to, but using it will make their life easier, as they will no longer need to maintain the JSON themselves, and what they build is the same thing that Dalamud will consume, and will match what the buildbot will build.
```
NotAdam, line 217: i thought people are just configuring some shit in the xml which is used by CI? why is there a json?

text/0000-automated-build-and-submit-pipeline.md
philpax: I left it as a top-level folder to make it clear that there's a folder => channel mapping, but yeah that's not a bad idea. Will think about it.

text/0000-automated-build-and-submit-pipeline.md
philpax: Good question. I ported that over from the existing [JSON](https://github.com/Ottermandias/GatherBuddy/blob/main/GatherBuddy.json), but we can probably get rid of this. Only question is if we want the extra check to prevent deployment of a testing version into main; I don't really have any strong feelings about that.

text/0000-automated-build-and-submit-pipeline.md
philpax: 🥲 

We have a [JSON file that shares the name of the plugin](https://github.com/goatcorp/SamplePlugin/blob/master/Data/SamplePlugin.json) and it is populated with some basic information about the plugin. DalamudPackager takes that existing JSON, and combines it with information from the `csproj` regarding AssemblyVersion etc to create the final JSON, which looks like [this](https://github.com/Ottermandias/GatherBuddy/blob/main/GatherBuddy.json).

That final JSON is then manually uploaded, alongside the zip of the package, to `DalamudPlugins`. I find this to be pretty redundant (why is there a step to combine the JSON and csproj into another JSON instead of just sourcing all the information from csproj?), but people have built workflows around this and some people manually maintain the final JSON themselves...

There's a bit of a general reticence to use `DalamudPackager` as well. @Eternita-S might be able to talk about their reasons/workflow here.

NotAdam: yeah okay, so the easy solution is to not use dalamudpackager and do it in a less shit way

sounds good to me 👍

kalilistic:
Should probably add @daemitus to this one too given his work on the existing github action.

KazWolfe:
As part of this process, we should add Actions to the PR process to verify that:

1. The plugin actually builds properly.
2. The resulting manifest contains all required fields (e.g. name, punchline, version)
3. The plugin passes a basic automated test suite (containing what? Signature validation? Security?)

karashiiro:
Oh yeah, I'm not sure if it's useful or not but I just remembered I had this <https://github.com/karashiiro/Dalamud-Checksummer/blob/main/Dalamud-Checksummer.ps1> that I was working on a while back.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+With the initial submission, proposed updates or final deployment, the GitHub Action will pull down the `repository` and build the `csproj` located within `location` (or the root) at the specified `commit`. This code is then automatically packaged with DalamudPackager, with the `pluginname.json` being generated using metadata drawn from the `csproj`, and this is used to produce an artifact. This artifact is then made available for download, and upon merging of the PR, the artifact is deployed to the `DalamudPlugins` repository. This build process could be later extended to feature additional verification steps (see [future-possibilities]).
+
+The primary role of the action is to consume a plugin manifest, and produce an artifact that can be loaded by Dalamud, or be included within `DalamudPlugins`. `DalamudPlugins` will only be updated on deployment (e.g. committing of the manifest file to the `master` branch in the repository); it will not be updated otherwise, so there is no risk of the CI accidentally committing an incomplete plugin to the masterlist.
```
ayyaruq, line 95: This is predicated on your `DalamudPlugins` repo being a good source of truth, which imo isn't necessarily the case, especially with multiple release streams. I think this also isn't clear with *where* the GHA build job runs. If it's in a single repo, the org will run into resource limits, and if it's not then you need all the overhead of setting up GHA to make PRs to other repos and handle crap like merging for conflicts.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+The `csproj` file should describe how to build the plugin. There may be cursory automated checking of the `csproj` to ensure that it is not doing anything malicious, but the primary verification strategy will still be human review. One of the checks will be to ensure that there are no `DalamudPackager` steps that will run on the CI itself, as to prevent double-packaging.
```
ayyaruq, line 97: You could likely use a bot for this. Something like PR goes up, it goes into a state of waiting for project config and manifest review, then an approved reviewer can post a comment like `/init-review pass` that the bot then triggers the GHA for and updates labels for state. Kinda meme, but there's enough open source things like this to quickly throw something together.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+Developers are also encouraged to ensure that the `csproj` is only responsible for building the plugin, and does not build anything unrelated. This may require developers to partition their project, or use a conditional build step similar to the one described above. Additionally, if the plugin has a build-time dependency on something, it is best if that dependency is built ahead of time and the result committed to the repository, as the CI will not be able to build the dependency.
+
+To encourage CI compatibility, as well as making it easier to test, a goatcorp-blessed GitHub Action will be made available that plugin developers can include as part of their repository. This action will build the package, similarly to how the CI would do it, so that developers can test locally with [act](https://github.com/nektos/act) or similar.
```
ayyaruq, line 118: I'd also update the various template repos to include this, and make sure they're actual GH templates. Prob mainly a note for @karashiiro if this goes ahead.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Rationale
+
+If we do not implement a solution like this, plugin submission and review will remain tedious. This will cause an increasing amount of frustration for everyone involved, especially as we continue to attract new plugin developers and thus more plugins and updates. Each step in the packaging process is a potential failure point, and supporting a developer through a failure point takes valuable time away. By removing this friction, we can make the experience better for both developers and goatcorp.
+
+This has downstream effects for users, as:
+
+- **the turnaround time for plugin updates will be faster** due to the process being significantly simpler. A developer can make a change, commit it, and then use the GitHub web UI to open a PR to update the manifest; a goatcorp reviewer can then look at the change and approve it, all without having to leave their browser.
```
ayyaruq, line 138: I don't think making a PR in the UI is a benefit, but it's also not required here. Could always have a separate goatcorp GHA in user-repos for GH users to trigger on new tags/releases and auto generate the PR. This also allows using tags which is generally a better idea than raw commit refs, provided they're not repointed.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+- **new plugins can be reviewed faster**, as it will be easier to ensure that the developer is not doing anything malicious, and that what reviewers see through the web frontend is what is being submitted (which is presently not a guarantee).
+- **users can be confident that plugins have been audited to some degree**; while most users do not directly care about this, someone _should_ care, because deploying malware to 100k people would suck
```
ayyaruq, line 140: This is where a notary service is of value, it doesn't affect the security but it *does* verify the supply chain which is of value. You also get an audit trail of who reviewed what, and easier revocations.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+## This design
+
+This design has a few key benefits over competing designs:
+
+- We don't pay for compute
+- It is fully declarative (e.g. what's specified in the metadata repo is what's deployed)
+- It does not require us to trust plugin developers' build systems
+- It is not a significant change from our existing system
+
+Initial discussion proposed a web portal, likely in addition to existing goatcorp services, where developers could submit their plugins to be reviewed. While this would be nice to have, it's not really immediately necessary.
```
ayyaruq, line 152: There's value in a web portal, but this specific way of centralising things makes it much less useful. A more appropriate use of a web portal in this proposal would be for indexing the downloadable artefacts instead of users needing to hit up a git repo for the source of truth, they can't see the git stuff anyway in XL but it gives you a static source relative to one that can be manipulated.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+#### Removing a plugin
+
+- The developer opens a PR to remove the relevant manifest from the channel
+  - We'll want to consider the interaction with [Dalamud#814](https://github.com/goatcorp/Dalamud/issues/814), which provides more information about why a specific plugin is no longer available
+- Upon merge, the action will detect this and update DalamudPlugins appropriately
+
+## Alternatives
+
+An alternative design was proposed by [Ay](https://github.com/ayyaruq) where plugin developers, upon main repo approval, would be given keys exclusive to the plugin that could be used to sign future updates. This would allow new versions of plugins to be deployed without explicit goatcorp approval. However, this solution was rejected due to its relative complexity for limited payoff.
```
ayyaruq, line 201: The main bulk of my suggestion was actually notarising, not signing. Code signing says a plugin is built by a goatcorp trusted developer, it says nothing about the plugin or build process being safe. Notarising gives you an audit trail of who specifically did what so you can see the full data provenance of the artefact, *and you can revoke ones found to be bad after release* which is the important bit. While I do think some bits of this are more complex than this proposal, I think the workflow is simpler and the infra needed for notarising isn't particularly complex, hell I'd built it myself if wanted.

The key issue here is what the goal is. If you want to know a plugin has been built by an approved dev, code signing works. If you want a nicer build process, then the security components aren't really relevant. If you want to improve user trust, notarising works. Different solutions for different goals.

text/0000-automated-build-and-submit-pipeline.md
ayyaruq: Using the name of the plugin in the manifest file is a bit iffy, just roll with `manifest.json` or whatever and be done with it. A consistent name means less faffing around with config handling and easier for newbie devs reading other repos for ideas.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+- <https://discord.com/channels/581875019861328007/860813266468732938/972255490128113684>
+  - Concerns raised during this discussion:
+    - **Not every dev wants to use DalamudPackager**: they won't need to, but using it will make their life easier, as they will no longer need to maintain the JSON themselves, and what they build is the same thing that Dalamud will consume, and will match what the buildbot will build.
+    - **How will the Dalamud community benefit from these changes?** See [rationale-and-alternatives].
+    - **Will this lock out projects with unconventional build processes?** Hopefully not; as long as the buildbot can build _a_ `csproj` and produce a DalamudPackager artifact at the end, it'll be happy. Other projects or steps are fine, just make sure they're not in the build path.
+    - **What about build-time dependencies that I rely on, but don't want goatcorp to build?** We have no immediate plans to support this. Unfortunately, this means our current answer is "build the dependency separately, and commit it into the repository to be picked up by the buildbot."
```
ayyaruq, line 220: Or... just have a separate step to pull them down?

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+This would form the backbone of any future CI we do around plugin submission. Ideas for improvement include:
+
+- Implementing a web frontend for the plugin submission/review/update process. Instead of manually PRing up changes to the metadata repo, a plugin developer could pop over to the frontend and specify which commit they'd like deployed, then hit the button.
+- As we're building the code, we have an opportunity to analyse it and ensure that it isn't doing anything outright naughty. We would likely use a solution similar to [S&box](https://wiki.facepunch.com/sbox/AccessList)'s. This could be combined with other steps, like asking the plugin to declare the namespaces it will use.
+  - This is pretty hard to verify when unsafe code is enabled, as all bets are off. This would be a relatively coarse check to make sure that there isn't anything _immediately_ suspect.
```
ayyaruq, line 230: Realistically this will require some custom static analysis tooling. Not sure how hard this is to do, but it's likely the only way.

ayyaruq (commented):
Overall looks good, I have some nits with the user experience, and I think there's over-emphasis on things that are't fully defined. There's also some customisation decisions that don't make sense for a build service.

While I agree that a decentralised approach makes the most sense, I'm not sure forcing everything into a goatcorp repo is the way to do that. Consider a GHA plugin that can run directly in user repos and submits a manifest for review. This way the free build hours aren't shared across all users, you get more build concurrency, and if the actions are well made you can still enforce all the checking.

A separate "CD" process or whatever could then check the actual code before making the manifest public to users. I'd likely proxy the URLs to GH with a perma redirect tho, so users hit the CF fronted web server for the manifest and then download it from GH releases instead of wasting goatcorp bandwidth. The downside here is your GHA's need to be better, and it's slightly more iffy for non-GH users that would need some goatcorp repo to use instead. I expect some hybrid is likely the actual answer, or otherwise centralise it and try find a way to make builds not cost a fortune. More repos is likely not the answer tho, especially if it involves manually PRing what's essentially a version string bump. Just automate the PR outright.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+Each folder contains files that correspond to a submitted plugin, which is what plugin developers will submit and update. The file will follow this structure (TOML):
+
+```toml
+[plugin]
+# The open-source Git repository that your code is hosted on.
+# Must be visible to the external web, but doesn't have to be GitHub.
+repository = "https://github.com/Philpax/plogonscript.git"
+# The commit to check out and build from the repository.
+commit = "9ef08248497c7f9f1312e14c50650183ee365734882a655345fc06abdac511f1"
```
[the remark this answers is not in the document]
KazWolfe: I'd tend to agree, but tags and branches are immutable. If we want to trace the lifecycle of plugin code to a specific and exact version, I think we'd have to use commits.

Otherwise, a malicious actor could create a tag `v1.0.1` with malicious code, wait for it to pass through review, and then force-push/retag to a clean code version. I'm unsure how much of a concern this would *realistically* be, but it feels like it could create a bit of a headache.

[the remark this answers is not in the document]
ayyaruq: I don't think that's much of an issue. When new commits are pushed, GH will require updated approvals. As far as CI goes, you grab the ref from the tag so if the tag changes, so does the ref and it's no longer what was approved.

kalilistic (approved):

ayyaruq:
> I'll make a note about making sure the action is compatible for local running...

Generally stuff that relies on GitHub specific APIs (like cache) doesn't work in `act`, anything else should be gucci.

kalilistic:
Maybe I got excited and approved early. Looks like this is pending some cleanup by @philpax based on the additional comments.

text/0000-automated-build-and-submit-pipeline.md
philpax: So, what we have now is a two-variant process:
- an initial `pluginname.json` that the dev fills with some basic information about their plugin
- DalamudPackager, or a very picky dev, then fills in that `pluginname.json` with the rest of the required information, most of which is sourced from the `csproj`

I think it's a bit silly that we have two versions of the same file, especially when the latter is much more clearly meant to be the result of an automated process (except, again, for those very picky devs with their own workflow)

That's why I'm thinking we can move the actual user-specified properties to the `csproj`, and then pull from that to generate the final manifest. Naming it `manifest.json` is totally fine by me, just depends on how much rework we do of the plugin loader.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+Upon completion, this should allow for developers to easily submit new plugins for review, check on the status of their plugin, and release updates, without having to build and package their own plugins. On the goatcorp side, it should ease the review process for initial plugin submission, reduce the overhead of reviewing new code, and remove the need to decompile incoming plugins (as what we receive is what we build is what we publish).
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+A new GitHub repository, tentatively `DalamudPluginManifests`, will be created. In this repository, there are several folders, with each folder corresponding to a different "channel" that may be built differently or be located elsewhere. At the time, these are the proposed channels:
+
+- `main`: The primary repository for Dalamud plugins, and what first-party plugin users will have access to. This is the most stringest to submit to, as it has the widest audience, but we should offer all the tools we can to reduce the barrier to entry.
```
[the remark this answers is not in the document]
philpax: lmao how did I do "stringest"? this is what happens when you write 4am DIPs. no opinion on the channel name, will change to `stable` unless anyone has very strong opinions

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+- `testing`: A repository for plugins that still require additional testing, or don't meet first-party requirements yet. The requirements to submit to `testing` are lower than that of `main`, but plugins still have to follow the general guidelines (e.g. no bots!)
+  - Additionally, to assist in the testing process, it is likely that update (not submission) PRs to `testing` will automatically be approved and merged. This allows developers to rapidly test changes, but ensures that human oversight will still be in place for when they submit to `main`.
+- `net6`: Plugins compiled for the .NET 6 version of Dalamud, so that they can be tested for the migration.
+
+Each folder contains files that correspond to a submitted plugin, which is what plugin developers will submit and update. The file will follow this structure (TOML):
```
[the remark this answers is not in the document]
philpax: I like TOML and think it's a perfectly acceptable fit for a currently-user-targeted configuration file, especially because it vastly constrains what's possible compared to YAML. JSON's _fine_ but is marginally more annoying to manually update.

Will stick with TOML unless someone else tells me that they simply cannot go on if it isn't JSON.

text/0000-automated-build-and-submit-pipeline.md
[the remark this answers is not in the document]
philpax: I'm with Kaz here - I think it's much easier to reason about what's actually deployed if you have a fixed version being specified. I think the benefit of being able to look at a manifest and knowing exactly which version is in play outweighs someone PEBKACing an out-of-branch ref.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+```toml
+[plugin]
+# The open-source Git repository that your code is hosted on.
+# Must be visible to the external web, but doesn't have to be GitHub.
+repository = "https://github.com/Philpax/plogonscript.git"
+# The commit to check out and build from the repository.
+commit = "9ef08248497c7f9f1312e14c50650183ee365734882a655345fc06abdac511f1"
+# Optional: where your project/csproj is located within the repository.
+# If empty or unspecified, it will be assumed the project is in the root of the repository.
+location = "plugin"
```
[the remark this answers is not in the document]
philpax: Did think about this for a very brief moment and thought that `location` was nice and terse, but yeah you're probably right, will switch to `project_path`.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+# The changelog for this version. Will be shown in-game, as well on the Goat Place Discord.
+changelog = "added Herobrine"
+```
+
+For this to work, the repository must be structured to be amenable to automatic builds and packaging, and it must be accessible to the CI pipeline (which in turn means it must be accessible to the public). Details of this will be covered in [reference-level-explanation].
+
+This DIP also proposes moving all properties associated with `pluginname.json` into the project's `csproj`, and having the `pluginname.json` be generated by DalamudPackager using the information within the `csproj`. Developers would still be able to maintain their own `json` if they really want to, but the new process should be easier (only one place to update all project-related metadata).
```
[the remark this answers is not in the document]
philpax: > Realistically most plugins do use a csproj, but definitely not all. 
What would you ("you" as in mystery plogon dev) use instead, and why? Is a non-standard configuration like that worth supporting?

`manifest.json` is fine by me though.

text/0000-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,231 @@
...
+
+```xml
+<ProjectExtensions>
+    <DalamudPlugin>
+        <ApiLevel>6</ApiLevel>
+        <Author>Ottermandias</Author>
+        <Name>GatherBuddy</Name>
+        <Punchline>Simplify Gathering and Fishing.</Punchline>
+        <Description>Adds commands to simplify gathering by finding nodes and fish and their locations via item name and a UI to keep track of special uptime and weather conditions.</Description>
+        <IconUrl>https://raw.githubusercontent.com/Ottermandias/GatherBuddy/main/images/icon.png</IconUrl>
```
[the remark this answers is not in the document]
philpax: Not sure what the current situation with that is - probably worth doing, but I think that's out of scope for this DIP

text/0000-automated-build-and-submit-pipeline.md
philpax: > This is predicated on your DalamudPlugins repo being a good source of truth, which imo isn't necessarily the case, especially with multiple release streams. 

If it _isn't_ a good source of truth, what's the point of doing all of this?

> I think this also isn't clear with where the GHA build job runs. If it's in a single repo, the org will run into resource limits, and if it's not then you need all the overhead of setting up GHA to make PRs to other repos and handle crap like merging for conflicts.

Fair point. I thought it was reasonably clear that we'd be the ones building, either through GHA or our own build server, but I'll tweak the language to make that a bit more obvious once this round of feedback's locked in.

I'm still rather attached to building the plugins "ourselves"; I think it's conceptually the simplest model for both goatcorp and developers to get their heads around, and it gives us free reign in doing whatever kind of validation / post-processing that we want to do for first-party deployment. I don't think chipping in for unrestricted builds would be problematic if it got to that point.

text/0000-automated-build-and-submit-pipeline.md
philpax: Yeah definitely, I'm most familiar with https://github.com/bors-ng/bors-ng myself

text/0000-automated-build-and-submit-pipeline.md
philpax: Agree to disagree on the commit refs, as before. And yeah, plenty of automated solutions for this; I just wanted to make it clear that updating your plugin would end up being _significantly_ simpler than the current DalamudPackager-and-upload dance, and is something you could do from your phone if you were so inclined

text/0000-automated-build-and-submit-pipeline.md
philpax: True, but my gut feeling is that it's far too much effort for both developers and us compared to just building their code ourselves

text/0000-automated-build-and-submit-pipeline.md
philpax: Yeah, a web portal's not super useful with the strategy described in this DIP. That being said, I refer to my previous comment - if we can't trust `DalamudPlugins`, who can we trust, and why are we doing this?

text/0000-automated-build-and-submit-pipeline.md
philpax: Yes, sorry, I've the unfortunate tendency to mix those up. I'll update the wording on the next pass and try to better communicate what you were proposing.

text/0000-automated-build-and-submit-pipeline.md
philpax: So this is in reference to something @KazWolfe brought up - XIVDeck has a JavaScript component that's built to be deployed to the Stream Deck, and may be provided by the plugin, but it's not part of the plugin itself. The concrete concern is that they did not want goatcorp to build the JS (which is in the same repo) because it's out of our purview, but at the same time, they wanted that JS to be available as a consumable artifact for the plugin.

Given that that's kind of annoying to support, I'm handballing it here by saying "build it yourself ahead of time within the repo". We can figure something out later if it ends up being more of a problem.

text/0000-automated-build-and-submit-pipeline.md
philpax: It's not too bad - @goaaats had a look at how S&box's solution works, and it basically just runs Roslyn over the codebase and scans over the resulting semantic information for naughties. Can add this whenever, I think

text/0000-automated-build-and-submit-pipeline.md
goaaats: I like the `testing/branch` scheme, makes sense to me

text/0000-automated-build-and-submit-pipeline.md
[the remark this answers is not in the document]
goaaats: I don't mind it either way - if we go with yaml, we have automatic support for json. I'm fine with TOML as well. In the end, the serializer we use probably supports both so I don't see why we can't offer both down the road, might just get annoying re: linting

text/0000-automated-build-and-submit-pipeline.md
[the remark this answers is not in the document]
goaaats: At the moment:
* Plugins on the main repo are ALWAYS loaded from the GitHub repo, the links in the manifest are completely ignored
* Plugins on the third party repo have their icons loaded from the manifest-provided links

This was a compromise we agreed upon when implementing this feature. I definitely want to avoid having regular Dalamud users leak accesses/their IPs to plugin-specified servers just by scrolling the installer, be it by having a media proxy or doing what we do right now(preferably).

text/0000-automated-build-and-submit-pipeline.md
goaaats: Note that checking for actual `unsafe` blocks is a pain in the ass, the only way to do it reliably is at compile time where you just don't turn them on. That would mean compiling it ourself, but then you can just pull in unsafe libraries from nuget.

I don't think we should worry about this right now. All discussion of security models for us has always resulted in too many limitations on plugin devs, e.g. no nuget, no unsafe, limited std lib - I think for now, without a lot of work, that ship has sort of sailed.

This DIP should first and foremost be about reviewability, IMO - the plugin team should be able to make sure that the ref specified is what produces the binary, being able to track changes, with an additional focus on dev experience.

text/0000-automated-build-and-submit-pipeline.md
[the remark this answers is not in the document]
goaaats: `pluginname.json` is stupid and should go, so `manifest.json` is fine by me.

philpax:
Just updated with latest round of feedback. Instituting informal RFC closure in a week's time, so the 31st. Get your feedback in now!

text/0000-automated-build-and-submit-pipeline.md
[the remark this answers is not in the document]
KazWolfe: Another vote in favor of keeping things as-is; icons being stored alongside the rest of the "app store" entry makes sense for mainline plugins. 

I'm not a huge fan of hardcoding different behavior for the mainline repo though -- would it be possible to always read `IconUrl` but then have the mainline `pluginmaster.json` always point to our own GitHub-controlled assets? We could then alter the build-time behavior to either download the image from the plugin-provided IconUrl or use the embedded icon at a well-known path.

Third-party repos are third-party repos; there's no way to really prevent IP leakage there without us *also* media-proxying actual plugins which opens a whole can of worms.

goaaats (approved):

karashiiro:
DIP number in the filename? Looks like it's still 0000.

philpax:
goat raised the very important issue of ensuring that only the right people get to submit manifest updates for their plugins. Still discussing best way to do that, but no-merge until that's in the DIP...

NotAdam:
just pull the userid from the gh api?

philpax:
> just pull the userid from the gh api?

How do you know which user ids are valid, especially for multi-contributor projects?

Anyhow, the solution I've settled on for now is to let people specify usernames for owners in the manifest they submit, and the pre-merge hook will just replace the usernames with IDs + username comment. Should handle changing usernames + future access control relatively easily, and we can always change it if it doesn't work.

Just need to write it up...

text/17-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,260 @@
...
+[plugin]
+# The open-source Git repository that your code is hosted on.
+# Must be visible to the external web, but doesn't have to be GitHub.
+repository = "https://github.com/Philpax/plogonscript.git"
+# The commit to check out and build from the repository.
+commit = "9ef08248497c7f9f1312e14c50650183ee365734882a655345fc06abdac511f1"
+# The people authorised to update this manifest (e.g. who can deploy updates).
+# Usernames can be specified here, but will be automatically converted to GitHub user IDs.
+owners = [
+    707827, // Philpax
```
ascclemens, line 49: This is the smallest nit possible, but if this is going to be a guide, it should probably have the proper comment syntax. `#` instead of `//`.

text/17-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,260 @@
+- Feature Name: Automated Build and Submit Pipeline
+- Start Date: 2022-05-03
+- DIP PR: [goatcorp/DIPs#17](https://github.com/goatcorp/DIPs/pull/17)
+- Repo-Relevant Issue: [goatcorp/dalamud#0000](https://github.com/goatcorp/dalamud/issues/0000)
```
NotNite, line 4: As I said on #18, are we keeping this or removing it (Repo-Relevant Issue)?

text/17-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,260 @@
...
+owners = [
+    "Philpax"
+]
+```
+
+but they will be automatically converted to GitHub user IDs before being merged in:
+
+```toml
+owners = [
+    707827, // Philpax
```
NotNite, line 72: As Anna said above, comment here should also be `#`.

text/17-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,260 @@
...
+
+```xml
+<ProjectExtensions>
+    <DalamudPlugin>
+        <ApiLevel>6</ApiLevel>
+        <Author>Ottermandias</Author>
+        <Name>GatherBuddy</Name>
+        <Punchline>Simplify Gathering and Fishing.</Punchline>
+        <Description>Adds commands to simplify gathering by finding nodes and fish and their locations via item name and a UI to keep track of special uptime and weather conditions.</Description>
+        <IconUrl>https://raw.githubusercontent.com/Ottermandias/GatherBuddy/main/images/icon.png</IconUrl>
```
NotNite, line 90: Not sure where this was said (Discord?), are we going to whitelist this URL?

text/17-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,260 @@
...
+
+Finally, build successes and/or failures will be published to the #github Discord channel, or its equivalent, to keep plugin developers in the loop. This is not a strict requirement, and may be adjusted to ensure that the signal-to-noise ratio stays high.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+This will require all first-party plugin developers to abide by a common project structure that is amenable to being built by CI, and for the code of their plugins to be open-source. The latter is already true, but the former may be a difficult requirement, especially with some of the complex build arrangements that existing plugins use. Upon preliminary exploration, it seems like this should be possible, but we'll need to check in with more developers.
```
NotNite, line 148: Are the Sonar devs fine with this?

text/17-automated-build-and-submit-pipeline.md
philpax: I think the best play here is to create a milestone to track the issues/tasks this will spawn, and to update the link here to point to that milestone.

wolfcomp:
Just gonna quickly comment, is it possible to maybe extend the automated pipeline to be a sort of plug and compile GitHub CI/CD system instead of locked to a specific NuGet package is required. That would require a specific file output for the upload instead of a specific build system. This would possibly take a little longer to make, but would open up the gates a little for how many drawbacks there are going to be otherwise being locked to one specific NuGet package. 

*Making this comment because DalamudPackager errors on my system complaining about some files missing when it's not missing.*

text/17-automated-build-and-submit-pipeline.md
philpax: I think this is where that discussion happened: https://github.com/goatcorp/DIPs/pull/17#discussion_r868725530

Seems like we already have an OK solution here, but it's something we can revisit later.

lmcintyre:
> Just gonna quickly comment, is it possible to maybe extend the automated pipeline to be a sort of plug and compile GitHub CI/CD system instead of locked to a specific NuGet package is required. That would require a specific file output for the upload instead of a specific build system. This would possibly take a little longer to make, but would open up the gates a little for how many drawbacks there are going to be otherwise being locked to one specific NuGet package. 
> 
> *Making this comment because DalamudPackager errors on my system complaining about some files missing when it's not missing.*

I would assume this should be something to look into (more descriptive errors) in DalamudPackager rather than rethinking the process. There is definitely a file missing in your case - it's just extremely difficult to determine *where*, because the error message just says that it couldn't find a file.

wolfcomp:
> > Just gonna quickly comment, is it possible to maybe extend the automated pipeline to be a sort of plug and compile GitHub CI/CD system instead of locked to a specific NuGet package is required. That would require a specific file output for the upload instead of a specific build system. This would possibly take a little longer to make, but would open up the gates a little for how many drawbacks there are going to be otherwise being locked to one specific NuGet package.
> > _Making this comment because DalamudPackager errors on my system complaining about some files missing when it's not missing._
> 
> I would assume this should be something to look into (more descriptive errors) in DalamudPackager rather than rethinking the process. There is definitely a file missing in your case - it's just extremely difficult to determine _where_, because the error message just says that it couldn't find a file.
The error was in an outdated structure of [SamplePlugin](https://github.com/goatcorp/SamplePlugin/) that was fixed with [#12](https://github.com/goatcorp/SamplePlugin/pull/12) that I followed to how I set up the project.

philpax:
Yup, we sorted out Wolf's issue on the Discord (thanks for pointing it out!) Wolf's mentioned that they're happy with this now that their issue's been fixed 👍 

As I mentioned on the Discord, my preference is to make the happy path trouble-free than to provide ways off the path. Happy to revisit once we get some more experience, though!

text/17-automated-build-and-submit-pipeline.md
rreminy: As discussed here: https://discord.com/channels/581875019861328007/972311484782755870/980537973051375677

We been recently under heavy fire recently after we expanded to CN and I'd rather not open source Sonar as that will make it far easier for Sonar to be reverse engineered.

We banned the offending users however it was an arduous process as it was difficult to identify them and they were bypassing the only identifier at the time. We banned them by IP addresses but they were using multiple VPN services which made this process difficult. We banned multiple VPN services as a result. The process of identification has been made more robust since then, of which I cannot provide the details here.

I did point out as well that the build process is complicated as multiple binaries are built then merged. I can provide further details if needed. `SonarPlugin.dll` is the final target of the build process, being strictly essentially a loader as it takes the `AssemblyLoadContext` to load the embedded binaries and start/stop Sonar.

While it is true that its easy to decompile due to the nature of how .NET works, it still does provide a strong barrier for reverse engineering. Decompiled source has no comments or leads helping them understand how Sonar works on the inside. In addition, simply decompiling `SonarPlugin.dll` in itself will only get you the loader.

I'm alright with providing a secret to be used for the purpose of building Sonar with the new process, or continue being the exception for the time being.

philpax: Yep - up to goat for final approval, but my thoughts on this are that we should kick the can down the road for now and let Sonar continue to merge directly to DalamudPlugins for now. We can figure something out as we get more experience with the overall pipeline.

text/17-automated-build-and-submit-pipeline.md
```diff
@@ -0,0 +1,260 @@
...
+
+The project will be built with the version of Dalamud appropriate for the channel that is being submitted to. In most cases, this will be the latest publicly released Dalamud, but e.g. the `net6` channel may be built with a Dalamud with .NET6 compatibility, so that testers can easily test plugins for that branch. All other dependencies will be built or sourced from NuGet as appropriate.
+
+Developers will be encouraged to add a conditional step that will not be run on the CI to their `csproj` to run DalamudPackager. In this step, it will be running in a mode where it will output a generated manifest JSON and any files required for the plugin to an output folder, so that developers can easily test their changes and have an experience equivalent to the CI. They're also welcome to continue to build the full ZIP artifacts for second-party and third-party deployments. Note that this is not mandatory, as you can keep locally bundling the DLL and your own JSON if you _really_ want.
```
ascclemens, line 123: DalamudPackager will need to be changed if it's really meant that developers will be adding this step to their `csproj`. As can be seen [here](https://github.com/goatcorp/DalamudPackager/blob/fab64d4/DalamudPackager/build/DalamudPackager.targets), DP has default settings when the `targets` file is not present, as is common with dependencies that add build steps. If it's intended to change this behaviour, developers who rely on DP's default build steps will have to add them manually instead.

Furthermore, DP could instead have its defaults changed to automatically ignore CI builds, which is probably a better solution.

text/17-automated-build-and-submit-pipeline.md
rreminy: @goaaats Before the merge happens, do you approve to let Sonar merge directly to DalamudPlugins manually for the time being?

goaaats: @rreminy Yes, that's okay for now. We'll figure something out when the system is up and running, which will be quite a while out.

PR #18, opened 2022-05-03
Final Comment: Formal subteams

philpax:
Very rough proposal. Need to better select people for each subteam, and answer the unresolved questions.

PR #28, opened 2022-05-07
WIP: Plugin Installer Refresh

kalilistic:


daemitus:
wut

text/0000-plugin-installer-refresh.md
```diff
@@ -0,0 +1,86 @@
...
+
+The plugin installer code base is known to be messy due to significant enhancements being layered on overtime. This has likely introduced reluctance to touch the code and make further improvements. There are also some QOL features that could be introduced to make the plugin installer a better experience for our users and reduce the related support efforts.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The following code improvements will be made:
+- Simplify code (many redundant methods, classes, etc.)
+- Abstract out logic that has crept into the imgui windows overtime.
```
karashiiro, line 24: We should clarify exactly what this means, I envision the majority of the logic being pulled out into a PluginInstaller class somewhere else, which the UI would call into.

daemitus:
Its not really THAT convoluted, but it is largely the most complicated view we've got going on.

Abstracting more of the state into the manager is definitely a good idea, But governing everything thats going on is what the problem is. While a plugin is loading, freezing all the other buttons, handing load errors (and theres alot of variants), etc.

text/0000-plugin-installer-refresh.md
kalilistic: Yeah, exactly what I meant. I'll update to be more specific. There could be a few classes under the plugin manager to keep things organized too. Same idea as the existing category manager for example.

kalilistic:
Any other feedback on this before merging? Looks like no big feedback/concerns on this one. @philpax

philpax:
Looks fine to me, you've outlined what you need from it. I guess the only thing you really need is approval from the relevant team? we should probably sort out who's on that team first

karashiiro:
LGTM

karashiiro (approved):

fitzchivalrik:
Just to mention it, I think it also would be great if any UX/UI redesign keeps the possibility of controlling the UI with a controller in mind.  
Most of the time it 'just works' when using basic Dear Imgui components, but anything complex or custom has the possibility to break the controller flow in subtle ways or introduce surprising behaviour.

kalilistic:
Marking as blocked as I think this is ready pending the approvers on https://github.com/goatcorp/DIPs/pull/18.

karashiiro:
Not sure if this is considered a precedent, but [I think we're standardizing around not zero-padding DIP filenames](https://github.com/goatcorp/DIPs/pull/29#issuecomment-1120308013).

MidoriKami:
With a bit outcry regarding streaming plugins, more and more people have been requesting ways to hide plugins on streams. While there are ways to hide the ImGUI components, there's no way to hide things that are rendered to the Native Game UI.

What might help is some way for plugin creators to tag their plugins with something that indicates they render to the Native UI.

I believe Kal mentioned having a warning category for plugin jsons, this could potentially be used for this purpose, but it might be better to have something more specific to rendering.

goaaats (approved):

PunishedPineapple:
I'm not sure whether it fits under the scope of this DIP, but with other parts of Dalamud and the plugin system being localized (or localizable), would we want to add fields to the plugin manifest that allow for the installer to show localized punchline, description, etc. in the languages that Dalamud supports?

philpax:
Seems relatively within scope, but not sure what the situation is with the localisation of the rest of Dalamud / if we have any existing support for that.

goaaats:
Should be easy to do, I won't have capacity for it though any time soon

philpax:
This is now unblocked, but I don't think we can merge this yet; I think it needs more detail / diagrams. Ideally, it'd be in a place where someone with the necessary skills who's not part of the conversation at all can implement it. (mostly so that goat doesn't have to be the one to do it)

kalilistic:
> This is now unblocked, but I don't think we can merge this yet; I think it needs more detail / diagrams. Ideally, it'd be in a place where someone with the necessary skills who's not part of the conversation at all can implement it. (mostly so that goat doesn't have to be the one to do it)

Sounds fair. I think I've added everything I plan to. If someone wants to contribute diagrams they can fork my fork or add them here and I'll amend the original.

kalilistic:
Updated with notes on native UI modification and plugin metadata localization.

kalilistic:
Marking as "helped wanted" as this is blocked until we find someone interested in doing wireframes/mockups.

PR #29, opened 2022-05-07
WIP: Caprine Operator

karashiiro:
Related notes:
- @goaaats is hosting the service, I'm not sure if he finished setting it up yet, though. Users looking to try it out will need to wait until then.

### DIFF
--- a/text/0000-caprine-operator.md
+++ b/text/0000-caprine-operator.md
@@ -1,0 +1,189 @@
+- Feature Name: Caprine Operator
+- Start Date: 2022-05-07
+- DIP PR: [goatcorp/DIPs#0000](https://github.com/goatcorp/DIPs/pull/0000)
+- Repo-Relevant Issue: [goatcorp/DIPs#25](https://github.com/goatcorp/DIPs/issues/25)
+
+# Summary
+
+[summary]: #summary
+
+Caprine Operator is an email reporting service to be used for streamlining the human review process over repositories in the
+core Dalamud ecosystem. The Operator sends emails with aggregated repository updates to subscribers to reduce the need for
+manually checking a repository and figuring out what has been updated and what can be ignored. After the Operator has sent
+its first email to a subscriber, subsequent emails will only include updates since that point. If there have been no updates
+since the previous email, no additional emails will be sent to the subscriber.
+
+# Motivation
+
+[motivation]: #motivation
+
+The introduction of Caprine Operator should reduce the cognitive load on participating reviewers, allowing them to spend more
+energy focusing on other important projects. As an email-based system, it supports the processes of reviewers who have a robust
+email setup for directing their attention.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Currently (as a prototype), Caprine Operator supports reporting updates for the plugin repository. The plugin repository review
+process requires repository moderators to perform the following steps:
+1. Identify that a pull request has been opened.
+2. Determine if the pull request is an update to an existing plugin, or a new plugin.
+3. If the pull request is for a new plugin, add the `new plugin` label to the pull request.
+4. Perform a code review of the plugin, and add the `code reviewed` label to the pull request.
+5. Perform a rules review of the plugin, and add the `rules reviewed` label to the pull request.
+6. A core review team member will merge the plugin, provided that a majority of the core review team agrees to its addition.
+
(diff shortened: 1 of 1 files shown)


karashiiro:
I am also proud to have the honor of setting the precedent of having implemented the thing before writing the DIP, I'll let everyone else judge how prudent that was.

text/29-caprine-operator.md
```diff
@@ -0,0 +1,189 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Currently (as a prototype), Caprine Operator supports reporting updates for the plugin repository. The plugin repository review
+process requires repository moderators to perform the following steps:
+1. Identify that a pull request has been opened.
+2. Determine if the pull request is an update to an existing plugin, or a new plugin.
+3. If the pull request is for a new plugin, add the `new plugin` label to the pull request.
+4. Perform a code review of the plugin, and add the `code reviewed` label to the pull request.
```
kalilistic, line 33: I'd consider checking the following as part of the code checks:
- Check fields are set including name, internalname, version, repourl, punchline, icon
- Icon may not be included in all PRs so the check would be in PR or already in the directory
- Contents of the two json files match
- If testing plugin, ensure [Testing] is in title (maybe even add it?)
- Consider standardizing PR titles (e.g. [TESTING] Update PluginName v1.2.3)

karashiiro: Can you clarify what you mean by icons? Missing icons aren't inherently an issue.

karashiiro: I can add JSON validation to the reports, auto-correcting titles requires an API key though, and I'm not sure if permissions can be scoped appropriately on those. I was looking for a way to get a locked-down API key to dodge the API rate limits, and it doesn't look like there's any way to do that.

karashiiro:
@philpax Just checking, are the DIP numbers supposed to be padded with zeroes or not? Rust RFCs do that, but PEPs don't. I figured we would avoid padding so we can hypothetically go to five or more digits without needing to change everything.

philpax:
I haven't been too discerning about it, but no padding should be fine.

text/29-caprine-operator.md
kalilistic: > Can you clarify what you mean by icons? Missing icons aren't inherently an issue.

I was trying to find the discord discussion but no luck yet. I thought I remembered some discussion with goat where we were considering requiring icons/punchlines to help improve the installer. Maybe that's a good open question that others can weigh in on though.

kalilistic: > I can add JSON validation to the reports, auto-correcting titles requires an API key though, and I'm not sure if permissions can be scoped appropriately on those. I was looking for a way to get a locked-down API key to dodge the API rate limits, and it doesn't look like there's any way to do that.

I only figured it out using PATs myself which sucks. I think the title edits are a nice to have though. The JSON validation is the important piece to avoid issues.

ayyaruq:
I'd probably expand on the use of PR labels, and maybe provide alternatives to email. Overall this sounds like generic GH bots a lot of large projects use, except specific to email.

karashiiro:
It'd be a bit ironic, but I could add a Discord webhook to the Operator, which would show the same information but structured slightly differently.

karashiiro:
On second thought, I'll add that to the future work section of the DIP, it's a good idea but not crucial.

> Overall this sounds like generic GH bots a lot of large projects use, except specific to email.

Do you have any specific examples that I can reference in the document?

karashiiro:
I updated the DIP updated based on everyone's feedback.

@kalilistic what do you think about [this](https://github.com/goatcorp/DalamudPlugins/pull/2497)? There's an empty image field in the metadata, but apparently Dalamud filters those out, so it doesn't actually cause any problems. Should this trigger an Operator warning?

kalilistic:
> 

Yeah, the meta data fields are only used for 3PP so can probably safely ignore.

karashiiro:
Updated the Operator, is there anything else, then?

kalilistic:
> Updated the Operator, is there anything else, then?

Maybe overkill, but this is technically block I'd say until the proper approvers are confirmed in https://github.com/goatcorp/DIPs/pull/18.

philpax:
The Operator is up and running, what else needs to be done here? kinda hard because it happened the wrong way around 😄

karashiiro:
Nothing else needs to be done, the infrastructure's all up. This can be merged now that the linked PR has been merged already.

PR #43, opened 2022-07-30
DIP17: ditch GitHub IDs for usernames

NotNite:
This goes against what was described to harden account hijacking, but we saw implementation difficulties in Plogon. GitHub Actions do not provide the account ID, only the username, which would require extra work to convert.

`<goat> I changed the spec earlier because IDs hard`
`<goat> sorry phil`


rreminy:
I found this if it helps:

https://stackoverflow.com/a/17309026

example: https://api.github.com/users/rreminy

PR #47, opened 2022-08-19
Loggy

philpax:
Loggy is a viewer for tspack log bundles to make it easier to troubleshoot.

Stakeholder ping: @NotNite @goaaats @reiichi001 @goatcorp/community-moderation @KazWolfe

Once we have a better idea of what all of this should look like, I'll clean this up


NotNite:
The source of loggy is currently at https://github.com/notnite/loggy (can i move to goatcorp repo?). It's currently deployed at https://soulja-boy-told.me/loggy/ (should we move to GH pages?).

text/47-loggy.md
```diff
@@ -0,0 +1,140 @@
...
+    - generates highly-compressed tspacks containing all relevant logs
+    - try/catch littered all around places where things can be reasonably expected to throw
+        - it would be awesome if we wrapped hooks in try/catch but dunno how much we want to fork
+    - issues that dalamud can identify will lead to specific, custom exceptions
+        - "service not found" etc
+        - any information that loggy can use to diagnose should be supplied
+    - where possible, information should be structured and not require string parsing
+        - json structures
+        - nested exceptions should be represented as nested structures
+        - callstack should be an array
```
KazWolfe, line 61: Something to note here that may be a problem at some point: [`StackTrace` is a string.](https://docs.microsoft.com/en-us/dotnet/api/system.exception.stacktrace?view=net-6.0) 

While there are [decent examples](https://www.neolisk.blog/posts/2015-08-13-csharp-capture-full-stacktrace/) on how to split them up, these all basically go through splitting newlines. It may also be worth looking at [`System.Diagnostics.StackTrace`](https://stackoverflow.com/questions/40981420/get-the-actual-stacktrace-object-from-an-exception) to try to prettify things and get some better information. A precursory test reveals that this does give us quite a bit of information, but we'd need to make sure we're serializing things properly and that this method actually produces useful information for debug parsing.

philpax:
Can I get some confirmation that this is acceptable to @NotNite, @goaaats and @reiichi001 before I write it up formally?

NotNite:
LGTM

philpax:
This is done so there's not really much point keeping this alive. Only thing that might be nice to have is some documentation of how it all works but that can be done whenever

PR #59, opened 2023-01-02
DIP59: Lockdown Mode

Minoost:
[Rendered (Markdown)](https://github.com/goatcorp/DIPs/blob/8f59f9e956534a5fa1299953c0f83d379e7c6289/text/59-lockdown-mode.md)

### DIFF
--- a/text/59-lockdown-mode.md
+++ b/text/59-lockdown-mode.md
@@ -1,0 +1,86 @@
+- Feature Name: `lockdown-mode`
+- Start Date: 2023-01-02
+- DIP PR: [goatcorp/DIPs#59](https://github.com/goatcorp/DIPs/pull/59)
+- Repo-Relvant Issue: [goatcorp/DIPs#58](https://github.com/goatcorp/DIPs/issues/58)
+
+# Summary
+Introduce "Lockdown Mode" to Dalamud, a sandbox feature for Dalamud.
+
+# Motivation
+It's simply infeasible to manually review plugins won't do any harms to your system and currently one and only one slip is  enough to do harmful things on your system.
+
+By launching FFXIV inside an app container and limiting what processes inside the container can do, we can trust that even malicious plugins won't do anything bad to user's system.
+
+# Reference-level explanation
+On launch time:
+1. [Create a new app container](https://learn.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-createappcontainerprofile) with the id of `Dalamud.Container` and [derive the assigned SID](https://learn.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-deriveappcontainersidfromappcontainername) from it.
+2. Set [capabilities](https://learn.microsoft.com/en-us/windows/uwp/packaging/app-capability-declarations) and [DACL](https://learn.microsoft.com/en-us/windows/win32/secauthz/dacls-and-aces) on files appropriately. (See below)
+3. Launch `ffxiv_dx11.exe` using the SID derived from (1) to create the process it inside the app container.
+
+This process will be implemented on `Dalamud.Injector` as this is the process responsible for creating `ffxiv_dx11.exe`.
+
+## Capabilities
+Capabilities granted on the app container is as follows.
+
+| Capability | Reason |
+|------------|----------|
+| SECURITY_CAPABILITY_INTERNET_CLIENT_SERVER | To communicate with the game server and some plugins might act as a local server (not game related fyi) as well. |
+| SECURITY_CAPABILITY_PRIVATE_NETWORK_CLIENT_SERVER | Some plugins might communicate with devices attached to home network. |
+
+## File System Access
+DACLs we set on the file system to support the feature is as follows:
+
+> [!note]
+> As a reminder, `deny` takes precedence over `allow` policy. (e.g. If `user1` is granted rw access on `/foo/bar` but also denied write access on `/foo/bar` then `user1` won't be able to write on `/foo/bar`)
+
+| Path | Access Type | Outcome | Reason |
+|------|-------|-------------|---------|
+| $base_game | r-x | allow | This is self-explaintory. |
+| $game_config | rw- | allow | This is self-explaintory. |
+| $game_config/downloads | -w- | deny | This directory is where patch files are stored by the retail launcher and thus should deny write access to prevent sandbox escaping by dropping malicious .patch files. |
+| $xl | rwx | allow | Contains plugin modules(.dll), plugin configs, dalamud logs, etc. |
+| $xl/{addon, runtime, patches} | -w- | deny | Mostly same reason as $game_config/downloads except it's XL this time. |
+| $dalamud | r-x | allow | Normally this directory's policy is automatically inherited (CONTAINER_INHERIT_ACE) from $xl but it's possible that's not the case when people are building Dalamud manually. |
+
+Anything not listed here follows regular access check rules for appcontainers.
+
+# Drawbacks
+- Some plugins may not be compatible without changes.
+- Current implementation is incomplete. (See [[##Unresolved questions]])
+
+# Prior art
+- There have been attempts to limit the scope of what plugins could do. (See [#53](https://github.com/goatcorp/DIPs/issues/53))
+	- However, I don't believe hooking .NET APIs is the right approach as APIs you need to secure is astronomical and is extremely hard (if ever possible) to get it right.
+	- Not to mention .NET BCL already tried a similar thing with `AppDomain` then later it got abandoned and is replaced wih `AssemblyLoadContext`.
+
+# Unresolved questions
+Semi-working prototype("Proof of Concept") is available for testing on [here](https://github.com/goatcorp/Dalamud/pull/1048). Use [this plugin](https://github.com/Minoost/Dalamud.TotallyLegitPlugin) to launch a terminal and test whatever hypothesis you might have from there.
+
+Mind you it has few caveats. Notably:
+- `$base_game` directory is usually owned by an administrator with no ways to change the permssion as a normal user.
+	- This will be addressed eventually however is out of scope for this DIP. Just take the ownership of this folder for now.
+- A process created inside the container has low integrity level (low IL from now) by default.
+	- This can be seen as an additional layer of defense as low IL process won't be able to read from (if NO_READ_UP is set) or write to (if NO_WRITE_UP is set) files with higher IL.
+	- However, this also means that sandboxed ffxiv won't be able to write configs as files have NO_WRITE_UP by default.
+- What should we do about the screenshot directory?
+	- We had an idea where the launcher would parse `ScreenShotDir` then give write access to it but discarded as malicious abuse it to gain write access to any directory. (Malicious plugins can simply write into ffxiv config to get write access as they wish)
+- You can't connect to `localhost` from AppContainer. This means some plugins **will be incompatible** without some changes.
+	- This CAN be worked around using [NetworkIsolationSetAppContainerConfig](https://learn.microsoft.com/en-us/windows/win32/api/netfw/nf-netfw-networkisolationsetappcontainerconfig) function. (See [this blog post](https://googleprojectzero.blogspot.com/2021/08/understanding-network-access-windows-app.html) for more details)
+	- However, using this to allow loopback is out of scope for a MVP.
+
+To side step these issues this feature will only be available if `DALAMUD_EXPERIMENT_APP_CONTAINER` is set to `1`.
+
+# Future possibilities
+- Some issues (like screenshots, localhost mentioned above) might need help of special process (also known as a "broker") to properly solve it.
+	- A process running inside the container would ask the broker (through some kind of IPC channel) to do something sandboxed process can't do by itself.
+	- This is what modern web browsers do and is proven effective against a compromised process if it's already sandboxed.
+	- However, going with this approach is also no small amount of work. Does it really worth it? Why don't just tell user to use default location or PrtScr? (or your favorite screen capturing tools, really...)
+- Windows let you create another AppContainer inside an app container. (DinD, except this time it's called "Nested AppContainer")
+	- Can we leverage this to provide per-plugin isolations?
+
+# Helpful Resources
+- [Windows Internals, Part 1, 7th Ed.](https://learn.microsoft.com/en-us/sysinternals/resources/windows-internals)
+- https://learn.microsoft.com/en-us/windows/win32/secauthz/appcontainer-isolation
+- https://chromium.googlesource.com/chromium/src/+/HEAD/docs/design/sandbox.md#App-Container-low-box-token
+- https://wiki.mozilla.org/Security/Sandbox#File_System_Restrictions
+- https://googleprojectzero.blogspot.com/2021/08/understanding-network-access-windows-app.html


text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+
+Anything not listed here follows regular access check rules for appcontainers.
+
+# Drawbacks
+- Some plugins may not be compatible without changes.
+- Current implementation is incomplete. (See [[##Unresolved questions]])
+
+# Prior art
+- There have been attempts to limit the scope of what plugins could do. (See [#53](https://github.com/goatcorp/DIPs/issues/53))
+	- However, I don't believe hooking .NET APIs is the right approach as APIs you need to secure is astronomical and is extremely hard (if ever possible) to get it right.
```
goaaats, line 53: I agree, considering that MS is abandoning these APIs and telling people to containerize instead. Sadly, we don't have a great way to do that since plugins rely on being in-process, if I'm not missing some magic.

text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Unresolved questions
+Semi-working prototype("Proof of Concept") is available for testing on [here](https://github.com/goatcorp/Dalamud/pull/1048). Use [this plugin](https://github.com/Minoost/Dalamud.TotallyLegitPlugin) to launch a terminal and test whatever hypothesis you might have from there.
+
+Mind you it has few caveats. Notably:
+- `$base_game` directory is usually owned by an administrator with no ways to change the permssion as a normal user.
+	- This will be addressed eventually however is out of scope for this DIP. Just take the ownership of this folder for now.
+- A process created inside the container has low integrity level (low IL from now) by default.
+	- This can be seen as an additional layer of defense as low IL process won't be able to read from (if NO_READ_UP is set) or write to (if NO_WRITE_UP is set) files with higher IL.
+	- However, this also means that sandboxed ffxiv won't be able to write configs as files have NO_WRITE_UP by default.
```
goaaats, line 64: I think XL fixing this/starting Dalamud as admin for it to fix it when this is detected is sensible.

text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+- What should we do about the screenshot directory?
+	- We had an idea where the launcher would parse `ScreenShotDir` then give write access to it but discarded as malicious abuse it to gain write access to any directory. (Malicious plugins can simply write into ffxiv config to get write access as they wish)
```
goaaats, line 66: Specify directory in XL, override it with startup args for XIV to use? Then block changing in-game? Bit convoluted but seems like the only really sane option to me.

text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+- You can't connect to `localhost` from AppContainer. This means some plugins **will be incompatible** without some changes.
+	- This CAN be worked around using [NetworkIsolationSetAppContainerConfig](https://learn.microsoft.com/en-us/windows/win32/api/netfw/nf-netfw-networkisolationsetappcontainerconfig) function. (See [this blog post](https://googleprojectzero.blogspot.com/2021/08/understanding-network-access-windows-app.html) for more details)
+	- However, using this to allow loopback is out of scope for a MVP.
```
goaaats, line 69: IMO, we probably need this. There's at least 2 plugins on the main repo that open a socket on loopback - if not HTTP, maybe pipes? Whatever is easier, but I feel like a HTTP socket is not that big of an issue.

text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+
+To side step these issues this feature will only be available if `DALAMUD_EXPERIMENT_APP_CONTAINER` is set to `1`.
+
+# Future possibilities
+- Some issues (like screenshots, localhost mentioned above) might need help of special process (also known as a "broker") to properly solve it.
```
goaaats, line 74: Another question: Opening links in browsers, afaik there is an API to do this for UWP, like Android intents?

text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+	- A process running inside the container would ask the broker (through some kind of IPC channel) to do something sandboxed process can't do by itself.
+	- This is what modern web browsers do and is proven effective against a compromised process if it's already sandboxed.
+	- However, going with this approach is also no small amount of work. Does it really worth it? Why don't just tell user to use default location or PrtScr? (or your favorite screen capturing tools, really...)
+- Windows let you create another AppContainer inside an app container. (DinD, except this time it's called "Nested AppContainer")
```
goaaats, line 78: There is one plugin in the official repo that relies on external processes(Browsingway) to spawn Chrome and transport frames back to the game. I honestly think that this should be out of scope, but if it's easy to allow plugins to start processes in the same app container... maybe?

goaaats (commented):
A big pending question for me is FS access - it would be nice to have the "broker process" handle FS grants that will allow plugins to request ACLs for other directories, based on an external UI handled by the broker process. This is important for e.g. Penumbra and creative plugins that allow for saving/loading arbitrary files.

Another option would be to just have pre-allocated storage folders for plugins(see iPhone file manager), and scope the Dalamud file-manager accordingly. This would be a bit annoying though and would likely not be an option for Penumbra, since it doesn't like long paths.

Out of scope?

text/59-lockdown-mode.md
KazWolfe: One of the devs checking in that uses localhost access for a port - this is essential to the functionality of my plugin, at least short of a full rewrite to use named pipes or similar. I'm opening a socket against `127.0.0.1`/`[::1]` on an arbitrary (user-defined) port above 1024. Assuming that localhost access (especially listens) are blocked or restricted, I will need some alternate way of getting data in/out of the game. To that end:

* What changes would be necessary to keep my ability to host this port? Or will I need to completely re-architect for compliance with DIP59?
* How will Lockdown Mode affect the ability to create or use named pipes and UNIX sockets?

If possible, I'd also like to get clarity on what `SECURITY_CAPABILITY_INTERNET_CLIENT_SERVER` and `SECURITY_CAPABILITY_PRIVATE_NETWORK_CLIENT_SERVER` actually _do_. As it stands, MSDN's documentation is rather unclear as to what these mean, so it's hard to figure out what these permissions grant (does `INTERNET_CLIENT_SERVER` allow listening to 0.0.0.0 and accepting connections from outside the local network? If so, we may want to disallow that...).

text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+| Path | Access Type | Outcome | Reason |
+|------|-------|-------------|---------|
+| $base_game | r-x | allow | This is self-explaintory. |
+| $game_config | rw- | allow | This is self-explaintory. |
+| $game_config/downloads | -w- | deny | This directory is where patch files are stored by the retail launcher and thus should deny write access to prevent sandbox escaping by dropping malicious .patch files. |
+| $xl | rwx | allow | Contains plugin modules(.dll), plugin configs, dalamud logs, etc. |
+| $xl/{addon, runtime, patches} | -w- | deny | Mostly same reason as $game_config/downloads except it's XL this time. |
+| $dalamud | r-x | allow | Normally this directory's policy is automatically inherited (CONTAINER_INHERIT_ACE) from $xl but it's possible that's not the case when people are building Dalamud manually. |
+
+Anything not listed here follows regular access check rules for appcontainers.
```
KazWolfe, line 45: How will this affect, say, symlinks to other folders? I know a non-zero number of users will symlink their character data folders to a place in cloud storage (Dropbox, OneDrive, etc.) in order to synchronize character information between multiple computers. 

As a general aside, this is something that we will need to decide if we want to support _at all_ or not.

KazWolfe, line 45: How does this affect registry usage? At present there are no (known) plugins that use registry, and there probably shouldn't ever be, but I think specifically denoting that this will break registry access (if it does?) would be beneficial so that it doesn't become a gotcha in the future.

text/59-lockdown-mode.md
KazWolfe: What kind of overhead would we be looking at for a broker, and what would the broker's limitations be? (And, more importantly, how will we prevent exploitation of the broker?)

Losing the ability to open links (especially to GitHub, Crowdin, issue pages, Discords, donation pages, etc.) would definitely be a frustration and would require we reconsider certain aspects.

Additionally: the game itself will open links in the user's default browser. I haven't looked at how the game itself does it, but will this interfere with that? (**Edit:** Under the current PR, web links in-game no longer open. See `ffxiv_dx11.exe+16DE30` under latest version at time of writing.)

text/59-lockdown-mode.md
```diff
@@ -0,0 +1,86 @@
...
+3. Launch `ffxiv_dx11.exe` using the SID derived from (1) to create the process it inside the app container.
+
+This process will be implemented on `Dalamud.Injector` as this is the process responsible for creating `ffxiv_dx11.exe`.
+
+## Capabilities
+Capabilities granted on the app container is as follows.
+
+| Capability | Reason |
+|------------|----------|
+| SECURITY_CAPABILITY_INTERNET_CLIENT_SERVER | To communicate with the game server and some plugins might act as a local server (not game related fyi) as well. |
```
KazWolfe, line 27: As alluded to in another comment, what exactly does this capability _do_? Or, more accurately, what's the difference between this and `SECURITY_CAPABILITY_INTERNET_CLIENT`?

Plugins generally (probably) shouldn't be exposing things to the world via server, so I'm not sure if the `_SERVER` side is necessary. Or maybe it is to allow any sort of listening to ports?

text/59-lockdown-mode.md
KazWolfe: Related: if a plugin needs "scratch area" (temp files), how should they go about setting that up? I assume we can't use `$TEMP` anymore with these changes, so will we just need to use our own plugin folders?

Fortunately I can't think of any plugin that needs a scratch area _yet_, but I think it's notable enough (and reasonable enough) to warrant at least bringing it up.

karashiiro: TextToTalk optionally uses the registry for TTS engine fixups (most of the Windows voices aren't available to the BCL synthesizer by default). I dispatch a separate process with a UAC admin prompt for it, but I'm not sure if that will let it jump the AppContainer or not.

karashiiro: I can just have the plugin ask the user to download and run it separately, though. TTT shouldn't be considered a blocker for this.

KazWolfe:
How would this affect the multiboxing mutexes currently used by the game? If I understand this subsystem correctly, the game uses a mutex to prevent launching multiple (>2?) copies of the game simultaneously. At least per [this article](https://scorpiosoftware.net/2019/01/15/fun-with-appcontainers/) it seems as though the mutex would be named differently and therefore not necessarily behave how we'd expect.

Would it be possible, then, for a user to launch more than two copies of the game at once through XL? Or two copies through XL _and_ two copies through the retail launcher? Do we care about this edge case?

I also feel that the link system _needs_ to be resolved somehow before we can move this to prod. The game itself calls `ShellExecuteA` for URLs (easy example: System Settings > Sound > Spatial Audio), which are broken in Lockdown Mode at present. I'm unsure how best to resolve this particular issue - would it make sense to hook `ShellExecuteA` or the calling method at `+16DE30`? Is it in general a good idea to (intentionally) break a game component and then subsequently patch it?

Minoost:
@goaaats @KazWolfe 

That's so many questions I have to answer at once so let's break it down into bite-size chunks. (hope I didn't miss anything important.)

> @goaaats
> Another option would be to just have pre-allocated storage folders for plugins(see iPhone file manager), and scope the Dalamud file-manager accordingly. This would be a bit annoying though and would likely not be an option for Penumbra, since it doesn't like long paths.
> 
> Out of scope?

I want to take this opportunity make something clear: per-plugin isolation is not the goal of this DIP. This DIP aims to protect the system (i.e. outside of FFXIV) from a hostile plugin. (which can theoretically contain something like a ransomware, for example.) However, it **won't** protect a plugin stealing your FFXIV session token, gils, items, etc.

> @goaaats
> A big pending question for me is FS access - it would be nice to have the "broker process" handle FS grants that will allow plugins to request ACLs for other directories, based on an external UI handled by the broker process. This is important for e.g. Penumbra and creative plugins that allow for saving/loading arbitrary files.
>
> @KazWolfe
> What kind of overhead would we be looking at for a broker, and what would the broker's limitations be? (And, more importantly, how will we prevent exploitation of the broker?)

In terms of runtime overhead it'd be pretty much negligible. Most of the time the broker would (1) read a message via IPC channel (2) and verify it's something allowed (3) then open a handle for them (4) and finally copy the handle into the sandboxed process. Actual I/O or heavy-duty operations would happen inside the sandboxed process.

Yes, because the broker is a privileged process (from the sandboxed process perspective) and acting as a gateway to the system "broker" is something you really make sure it can't be abused to escape the container. This generally means you really shouldn't do "dump anything what sandboxed processes can't do on the broker because we can" and anyone having such attitudes **seriously** shouldn't make changes to the broker.

However, whatever it may happen to the broker this is still a massive improvement over "everything is free real estate" approach. (actually countable # of IPC calls vs. virtually infinite)

> @KazWolfe
> Related: if a plugin needs "scratch area" (temp files), how should they go about setting that up? I assume we can't use `$TEMP` anymore with these changes, so will we just need to use our own plugin folders?
> 
> How does this affect registry usage? At present there are no (known) plugins that use registry, and there probably shouldn't ever be, but I think specifically denoting that this will break registry access (if it does?) would be beneficial so that it doesn't become a gotcha in the future.

Of course virtually unlimited access to registry is going away while running inside the AC.

I only talked about file usage but most kernel objects (including registry, named pipe, mutex, etc...) follows the same rule: (Assuming Windows already checked [IL](https://learn.microsoft.com/en-us/windows/win32/secauthz/mandatory-integrity-control)) If an object in question has an ACE that **explicitly** allows either AppContainer SID or `ALL APPLICATION PACKAGES` SID to access, then the access will be granted. If doesn't, then it will be denied.

It's important that apps can access whatever `ALL APPLICATION PACKAGES` SID can do. As otherwise you won't be able to load core Windows modules like DLLs inside `C:\Windows\System32`.

This rule also applies to the registry: Currently `Dalamud.Injector` doesn't touch DACLs on  `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer` but it has `ALL APPLICATIONS PACKAGES` ACE that explicitly allows read operation so you can read those and its subkeys.

Last thing I want to talk about is that each AppContainers gets their own private directory on `%localappdata%/Packages/$container_id/AC` (with full access to it) and certain directories (including `$TEMP`) are redirected into it.

Kernel object equivalent of this "private directory" (e.g. for mutexes) is located at `\Sessions\$session_id\AppContainerNamedObjects\$appcontainer_sid`.

If you're concerned with specific use cases then please tell me more about the scenarios.

> @KazWolfe
> As alluded to in another comment, what exactly does this capability _do_? Or, more accurately, what's the difference between this and `SECURITY_CAPABILITY_INTERNET_CLIENT`?
> 
> One of the devs checking in that uses localhost access for a port - this is essential to the functionality of my plugin, at least short of a full rewrite to use named pipes or similar. I'm opening a socket against `127.0.0.1`/`[::1]` on an arbitrary (user-defined) port above 1024. Assuming that localhost access (especially listens) are blocked or restricted, I will need some alternate way of getting data in/out of the game. To that end:
> 
> -   What changes would be necessary to keep my ability to host this port? Or will I need to completely re-architect for compliance with DIP59?
> -   How will Lockdown Mode affect the ability to create or use named pipes and UNIX sockets?
> 
> If possible, I'd also like to get clarity on what `SECURITY_CAPABILITY_INTERNET_CLIENT_SERVER` and `SECURITY_CAPABILITY_PRIVATE_NETWORK_CLIENT_SERVER` actually _do_. As it stands, MSDN's documentation is rather unclear as to what these mean, so it's hard to figure out what these permissions grant (does `INTERNET_CLIENT_SERVER` allow listening to 0.0.0.0 and accepting connections from outside the local network? If so, we may want to disallow that...).

It's very unfortunate that there are no official documents about AppContainer with sufficient details online. Afaik it's only in the book "Windows Internals" published by Microsoft themselves.

**However**, you can read a blog post (I already linked this in "Useful Resource" section) from Google Project Zero to see what's the difference between `SECURITY_CAPABILITY_INTERNET_CLIENT_SERVER` and `SECURITY_CAPABILITY_INTERNET_CLIENT` (and its home network variants) in details.

The gist is:
- If current network profile is `Public` then
	- `SECURITY_CAPABILITY_INTERNET_CLIENT` allows outgoing connection to `0.0.0.0 .. 255.255.255.255` (Effectively acting as allow-all)
	- `SECURITY_CAPABILITY_INTERNET_CLIENT_SERVER` allows incoming connection from `0.0.0.0 .. 255.255.255.255` (Effectively acting as allow-all)
- If current network profile is `Private` then
	- `SECURITY_CAPABILITY_PRIVATE_NETWORK_CLIENT` allows outgoing connection to ~~`0.0.0.0 .. 192.168.0.0 | 192.168.0.255 .. 224.0.0.0 | 239.255.255.255 .. 255.255.255.255`~~ (apparently this part is different depending on which network you're connected to).
	- `SECURITY_CAPABILITY_PRIVATE_NETWORK_CLIENT_SERVER` allows incoming connection from ~~`0.0.0.0 .. 192.168.0.0 | 192.168.0.255 .. 224.0.0.0 | 239.255.255.255 .. 255.255.255.255`~~ (again, different setup for each network).

(Funny that there are no differences between `*_INTERNET_*`and `*_PRIVATE_NETWORK_*` if it's set to public. It's as if "You're not at home anyway so whatever lol")

[The blog post](https://googleprojectzero.blogspot.com/2021/08/understanding-network-access-windows-app.html) has enough details so you should be able to check IPv6 rules by yourself. (I'm not going to post it in here.)

img


Access to `localhost` is always going to stinky as there's possibility that untrusted process can just connect to unauthenticated services[^1] then exploit its capability to escape the container. 

We (and of course, Microsoft) don't have a control over what's installed in user's system so AC just blanket bans on all `localhost` communications.

Currently we don't do anything about it and follows the default rule. (Ban all `localhost` communications[^2]) Later this might be changed but even then it will be "only if you turn this on this setting" kind of thing.

You CAN however, communicate with an external services[^3] via named pipes if the service in question explicitly allows it. (Add ACEs with `ALL APPLICATION PACKAGES` or specific package SID with appropriate permissions.) Just be extra careful as clients connect to the service can be actively hostile. (Just like any internet facing services.)

If the process inside AC is the host, then create it inside the private space I briefly mentioned above.

[^1]: Services that thinks it's safe from an attacker when binds a socket to `localhost` only.
[^2]: Actually, you can recv/send stuff from/to `localhost` if it's within same AC. Read the blog post from Google if you're interested.
[^3]: Any process live outside of the container, really.

KazWolfe:
Thank you, this all makes sense. I'll just tack the responses on to responses:

> Kernel object equivalent of this "private directory" (e.g. for mutexes) is located at \Sessions\$session_id\AppContainerNamedObjects\$appcontainer_sid.

I would like to see this tested to see how it affects multiboxing, as it sounds like it would at present. If this is something we're concerned about allowing, then we may need to have something in place elsewhere to ensure that no more than two instances of the game can be running at any given time. I'll defer to Goat as to whether this is an issue or not, but I feel like we shouldn't be allowing that here.

> Currently we don't do anything about it and follows the default rule. (Ban all localhost communications Later this might be changed but even then it will be "only if you turn this on this setting" kind of thing.

> If the process inside AC is the host, then create it inside the private space I briefly mentioned above.

This is what I was afraid of, and will require either a complete rearchitecture of my plugin _or_ the ability to read the status of whatever localhost comms checkbox we include (if we even decide to allow such a thing). As such, I'd like to request that if we move forward with this proposal, that _ample_ time be given for developers to manually enable and verify compatibility with lockdown mode before we start shipping this default for everyone.

Are paths to the private space(s) in any way predictable? In other words, if I create a named pipe from the game process, how will anything outside the game even know where to look? The same would apply, I suppose, to SIDs as well. You'd be correct that I'd like to restrict named pipes as much as I can, but I need to know the parameters of _what_ to restrict those to and how those would behave. 

Obviously, the breakage of XIVDeck should not be a blocker, as the overall security of the plugin ecosystem is more important than a single highly-specialized plugin, but I'd like to be as ready for these changes as possible especially given I have virtually zero understanding of AppContainer. If XIVDeck dies as a result of this, then so be it.

KazWolfe:
> I also feel that the link system needs to be resolved somehow before we can move this to prod. The game itself calls ShellExecuteA for URLs (easy example: System Settings > Sound > Spatial Audio), which are broken in Lockdown Mode at present. I'm unsure how best to resolve this particular issue - would it make sense to hook ShellExecuteA or the calling method at +16DE30? Is it in general a good idea to (intentionally) break a game component and then subsequently patch it?

This was discussed with Goat in the Discord and we came to the conclusion that ShellExecuteA would be hooked and handle links... however we'd otherwise handle that case for plugins. I'm still not particularly a fan of this solution, but I suppose it's unavoidable if we want to go down this path.

Minoost:
> I would like to see this tested to see how it affects multiboxing, as it sounds like it would at present. If this is something we're concerned about allowing, then we may need to have something in place elsewhere to ensure that no more than two instances of the game can be running at any given time. I'll defer to Goat as to whether this is an issue or not, but I feel like we shouldn't be allowing that here.

Untitled2

Because FFXIV creates a mutex on `Global` namespace (which of course the sandboxed process don't have access to and thus failing multibox checking), I've [patched out multibox checking code](https://github.com/goatcorp/Dalamud/pull/1048/files#diff-b87e3ab4e64b889df416c2ba88458f6cf3645dfe74c90b9a24ec1063a9933b91) for now.

Redirecting this to `Local` namespace and thus allowing 4 instances to run at most (2 for with AC + 2 for without AC) seems to be acceptable trade off.

> Are paths to the private space(s) in any way predictable? In other words, if I create a named pipe from the game process, how will anything outside the game even know where to look? The same would apply, I suppose, to SIDs as well. You'd be correct that I'd like to restrict named pipes as much as I can, but I need to know the parameters of what to restrict those to and how those would behave.

Let's say some plugin created a named pipe like this:

```csharp
var pipeServer = new NamedPipeServerStream(@"Local\HelloDalamudContainer");
// ...
// pipeServer.WaitForConnectionAsync()
// pipeServer.ReadAsync(someBuffer)
// ...
```

If a process outside the game wants to connect to this pipe, the process outside the game need to do something like this:
```csharp
// See https://aka.ms/new-console-template for more information

using static System.Console;
using System.ComponentModel;
using System.Diagnostics;
using Windows.Win32;
using Windows.Win32.Foundation;
using System.IO.Pipes;

Debug.Assert(OperatingSystem.IsOSPlatformVersionAtLeast("windows", 8));

// https://learn.microsoft.com/en-us/windows/win32/devnotes/getting-the-session-id-of-the-current-process
uint GetSessionId()
{
    var ok = PInvoke.ProcessIdToSessionId(PInvoke.GetCurrentProcessId(), out var sessionId);
    if (!ok)
        throw new Win32Exception();

    return sessionId;
}

string GetNamedObjectPath(string appContainerId)
{
    unsafe
    {
        PSID psid = default;
        Span<char> path = stackalloc char[260];

        fixed (char* pPath = path)
        {
            try
            {
                var hresult = PInvoke.DeriveAppContainerSidFromAppContainerName(appContainerId, out psid);
                hresult.ThrowOnFailure();

                var ok = PInvoke.GetAppContainerNamedObjectPath(null, psid, (uint) path.Length, pPath,
                    out var pathLength);
                if (!ok)
                    throw new Win32Exception();

                return new string(pPath);
            }
            finally
            {
                if ((void*) psid != null)
                    PInvoke.FreeSid(psid);
            }
        }
    }
}

string GetAppContainerSidString(string appContainerId)
{
    unsafe
    {
        PSID psid = default;
        PWSTR psidStr = default;

        try
        {
            var hresult = PInvoke.DeriveAppContainerSidFromAppContainerName(appContainerId, out psid);
            hresult.ThrowOnFailure();

            var ok = PInvoke.ConvertSidToStringSid(psid, out psidStr);
            if (!ok)
                throw new Win32Exception();

            return psidStr.AsSpan().ToString();
        }
        finally
        {
            if ((void*) psid != null)
                PInvoke.FreeSid(psid);

            if ((void*) psidStr != null)
                PInvoke.LocalFree((IntPtr) psidStr.Value);
        }
    }
}


// Main
var sessionId = GetSessionId();
var containerSid = GetAppContainerSidString("dalamud.container");
var objectPath = GetNamedObjectPath("dalamud.container");

WriteLine($"session id: {sessionId}");
WriteLine($"container sid (for debugging): {containerSid}");
WriteLine($"obj path: {objectPath}");

WriteLine($@"Inside of FFXIV:  Local\HelloDalamudContainer1234");
WriteLine($@"Outside of FFXIV: Sessions\{sessionId}\{objectPath}\HelloDalamudContainer1234");

var pipePath = $@"Sessions\{sessionId}\{objectPath}\HelloDalamudContainer1234";

using var pipe = new NamedPipeClientStream(pipePath);
pipe.Connect(/* timeout */);

// `NamedPipeClientStream` inherits `Stream`.
pipe.Write(new byte[] { 0x01, 0x02, 0x03, 0x04, 0x05 });
```
(maybe there's a better way to handle this...)

Untitled

> I also feel that the link system needs to be resolved somehow before we can move this to prod. The game itself calls ShellExecuteA for URLs (easy example: System Settings > Sound > Spatial Audio), which are broken in Lockdown Mode at present. I'm unsure how best to resolve this particular issue - would it make sense to hook ShellExecuteA or the calling method at +16DE30? Is it in general a good idea to (intentionally) break a game component and then subsequently patch it

I wonder how [Launcher.LaunchUriAsync][uwp] handles this. Given there are restrictions on what can be passed to `Uri`, maybe it's handled by the broker?

[uwp]: https://learn.microsoft.com/en-us/uwp/api/windows.system.launcher.launchuriasync?view=winrt-22621#windows-system-launcher-launchuriasync(windows-foundation-uri-windows-system-launcheroptions

KazWolfe:
Thank you for the sample code - I think that resolves my personal issue (although it will still involve a pain of rearchitecting everything, but oh well). Moving on:

> I wonder how `Launcher.LaunchUriAsync` handles this. Given there are restrictions on what can be passed to Uri, maybe it's handled by the broker?

This seems acceptable to me, provided we don't just actually use `LaunchUriAsync` (assuming we can?). We'd need to ensure that whatever we do remains compatible with Wine as well though. I'd also have questions about how the user experience of this will act (will users receive a pop-up asking to open a URL?), but again not particularly relevant immediately. For now I think the assumption that `ShellExecuteA` would only be used for URLs is valid, but we should probably have _something_ at least in the back of our minds in case Square decides to do more/launch extra processes.

> Redirecting this to Local namespace and thus allowing 4 instances to run at most (2 for with AC + 2 for without AC) seems to be acceptable trade off.

No objections to limiting this to four, just want this out in the open so others can discuss appropriately. We can also have other things handle this - e.g. the broker holds the (native) mutex open as long as the game is alive.

---

Overall, I support this proposition, though I have concerns about how much this breaks and for what purpose we're causing breakage. I'll always agree that more security is beneficial, but I do want to point out (what I see as) our risk model here: third party plugins and those that load in information from untrusted sources.

All first-party plugins are already reviewed and Lockdown Mode won't change that. This PR itself denotes this as a limitation: plugins can still be malicious in such a way that impacts or destroys a user's game. A plugin like that shouldn't ever really _be_ in the official repository anyways, and we can't stop that so long as plugins have full access to FFXIV's process space. It's true that our reviewers may not necessarily be able to spot suspicious or well-hidden security flaws/malware that would cause harm to a machine, but anything like that would already probably look odd or out of place in the context of an FFXIV plugin.

That leaves two notable threat surfaces that Lockdown Mode would protect against:

* **Malicious third-party plugins.**  
We've already gone on record and made it extremely clear that we cannot review or vet third party plugins in any way, shape, or form. They are untrusted executables running on a user's computer, and it is the user's obligation to ensure that _if_ they want to go into third-party space, they need to keep themselves safe. This is virtually identical to a lot of the existing security mindset out there: don't install things you don't trust. Granted, humans are, well, human and bad at making informed decisions, but I question the degree to which we should prevent users from footgunning themselves; especially as there are many many other points of entry for someone to grab malicious code (ACT plugins as well as full-on external executables like TexTools and Anamnesis).
* **Poorly-Written/Flawed Plugins or Systems**  
We already see crashes and issues caused by some plugins that load in invalid data (especially Penumbra mods). I suppose it would be possible for someone to create and distribute a mod that actually triggers an ACE due to a flaw in a plugin or the game itself. Lockdown mode would be an effective preventative measure here (as well as for similar plugins that may accept untrusted input). I'd agree that it's out of scope for plugin reviewers to try to find these cases, doubly so for the third party plugins that are currently the forefront of the threat surface for this point.

This DIP, overall, carries _a lot_ of potential to cause breaking changes. Some of these breaking changes are simple; e.g. changing file cache paths to more appropriate places or altering how certain behaviors work to better comply with a sandbox. Others may fundamentally require plugins rethink their approach: how/where to store and load files, how to perform inter-process communication, how to allow users to export data, and a whole host of other things. 

None of these are _bad_, necessarily, but I think we as a community need to decide whether these security benefits are worth the cost we'd have to pay in terms of rewriting plugins/supporting users during the transition/making certain documentation very clear on the user's responsibility to comply with the sandbox. Doubly so if we're doing this while a user can just download some malware in the form of an ACT plugin that promises to make your parses better. 

In short: if this becomes an enabled-by-default feature, are the security benefits material enough to outweigh the costs? We aren't a web browser where untrusted information exists in _vastly_ higher quantities; virtually everything done in the XIV plugin ecosystem is done with a degree of intent and (hopefully) foreknowledge as to the risks.

---

I think before moving forward, we should have a (documented) idea on how to solve a lot of the common problems:

* How will we communicate filesystem access requests to end users? What information will be provided to them, and how will it apply?
    * What will the process to grant/revoke consent to access things look like?
* How will this affect support, especially for the edge cases of users doing odd things?
    * How can we mitigate these support costs and confusion?
* How will we assist plugin developers with making their plugins compliant?
    * What changes should plugin devs (especially those that do a lot of file-based work like Penumbra) expect? How would a dev be able to tell if consent was granted or not?
* Are there going to be types of plugins that are just no longer going to be possible?
* What will the launch strategy of this feature be?
    * How long will the community have to adapt to changes? 
* What will we do if/when SE does things that are incompatible with the AppContainer?
    * Will there be any impact on other integrations (both official and unofficial) that want to tie into the game?

Again: all the underlying principles seem sound and these security improvements are not without merit. It just becomes a question of dragging kicking and screaming users/devs (like myself) into believing that the improvements are worth the support burden/change in how everything operates.

Minoost:
> This seems acceptable to me, provided we don't just actually use LaunchUriAsync (assuming we can?)

Oh wow, I haven't thought about that. `LaunchUriAsync ` works just fine.

(Not sure about )

https://user-images.githubusercontent.com/1381835/211057031-666ed517-3b24-4d51-9e2a-8b5eab4002c2.mp4

This does bring the minimum OS requirement to Windows 10 but [Windows 8.1 EoL](https://support.microsoft.com/en-us/windows/windows-8-1-support-will-end-on-january-10-2023-3cfd4cde-f611-496a-8057-923fba401e93) is just 3 days away from now so it's probably fine.

(I have no idea how it will work on Linux however..)

Minoost:
Looks like community's general consensus for this is "We don't need it" - ranging from legitimate concerns like ergonomic and maintainability issues to "we didn't get hacked before, so why bother?" and some downright aggressive and threaten to pull the plug if this feature were to be accepted. First round of such *discussion* apparently has happened **on early January** and in a private channel on Discord.

As such, I don't believe this needs be developed any further nor worth the time anymore.

As a side note, if anyone wants to keep working on this concept you're free to do so. Only task left to launch the game within the container is patching out/inserting shims for multi-box and `PROCESS_VM_WRITE` checks - that part should be easy enough.

PR #63, opened 2026-04-10
DIP63: Plugin Repo V2 Schema

KazWolfe:
Starting off DIP63. This PR will serve as the commentary and general discussion thread for what RepoV2 will look like.

Plugin developers and maintainers should use this thread to post commentary about what they want to see in RepoV2. As features/specifications are accepted, they will be added to the official DIP documentation, where they will be subject to further commentary/review.

To start discussion, @goaaats has provided [a draft](https://gist.github.com/goaaats/51b4c8fc7574539d295e786918a4998b) of what RepoV2 could look like.


63-repov2.md
```diff
@@ -0,0 +1,57 @@
...
+
+[motivation]: #motivation
+
+TBD
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+TBD
```
KazWolfe, line 21: Add comments about high-level features you want to see under this review comment, or on this line in the file.

63-repov2.md
```diff
@@ -0,0 +1,57 @@
...
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+TBD
```
KazWolfe, line 27: Add comments about specific implementation details you want to see here, or on this line in the file.

63-repov2.md
```diff
@@ -0,0 +1,57 @@
...
+
+[summary]: #summary
+
+Repo V2 is a new standardized repository format meant to clean up various quirks and problems that have accumulated throughout the years.
+
+# Motivation
+
+[motivation]: #motivation
+
+TBD
```
KazWolfe, line 15: Add complaints with existing repo system here, or on this line in the file.

63-repov2.md
KazWolfe: Brief things I personally want to see:

* Dalamud informing the server if a plugin download is an update or a fresh install.
  * Perhaps include the old version in an HTTP header?
* The ability to specify the minimum version of any given plugin that can load, e.g. disabling all versions before 1.2.3.
* Multiple branches for plugins, allowing users to select if they want to be on stable, beta, or any other such branch a developer may publish.
  * Allow setting branches as "hidden" so the user needs to type the branch name.
* A URL referencing a plugin's Git repo and current commit hash.
  * Optional URL for attestation metadata, if available.
* Optional fields for support URL, feedback destination, etc.
  * Feedback will POST a Dalamud-controlled JSON object to that URL. Recipients are responsible for transforming this accordingly.
* A field to specify optional repos that may also update this plugin, e.g. for repo rotation, devs who wish to publish bleeding-edge builds of their plugins, or custom-to-mainline pipelines.
* Some ability to localize plugin entries.

I'd also like RepoV2 to clean up the special edge cases that our existing repos have, especially with how icon and image URLs currently work for mainline.

KazWolfe: We also need to bear in mind that whatever we create needs to be statically hostable and simple to modify with standard shell tools (e.g. `jq`). However, certain *optional* fields and features may be added for users that wish to leverage more advanced API-like features.

63-repov2.md
KazWolfe: At present, the repo and manifest systems are very heavily intertwined. This is the first step to breaking these two apart. The repo should, ultimately, only be responsible for what's necessary to list in the plugin installer/perform updates, while the manifest will be responsible for runtime/installed data.

63-repov2.md
nathanctech: The custom feedback URL has been long requested as the current system isn't really feasible for a lot of maintainers who may not use Discord regularly. The feedback is also not directly actionable unless the person provides contact information which is rare, and I doubt many devs will resort to emailing people. We should definitely include this.

63-repov2.md
nathanctech: Discoverability in the current system is poor, and a good category system is a first step in that I think. We should also maybe think about giving Dalamud-specific categories for that purpose. Custom ones are great, but it can be difficult to search depending on how the developer specifies them.

63-repov2.md
rreminy: A custom feedback URL would be greatly useful.

I myself do look at the Discord from time to time *(although I'm not exactly active)*, however its difficult for me to follow the feedbacks channel due to the amount of "noise", since you'd have to look for feedback for your plugin along all the other feedbacks for all the other plugins, along with conversations that happen in between.

MKhayle: A badge showing whether AI was used in the process of making the plugin, whether assets or code, maybe? 
Not in a way to shame, but just to inform users about it. That'd need honesty from the developers themselves who would be the ones to assign it themselves (or when asked to by PAC, I guess).

MidoriKami: An AI Usage Policy is in the works, it essentially requires people submitting plugins/updates to disclose that they are using AI.

When looking at the code it's pretty obvious when AI is being used currently, and in places where it's not obvious normal code review practices still apply, so users can still feel confident that the merged plugins are good to go.

Regarding the badge, I almost wonder if its something that PAC should be able to apply to a plugin, say like via dalamud assets or some other mechanism, instead of having the dev themself note it in a manifest.

KazWolfe: Thinking about it: custom_feedback_url should probably be in the installed manifest, not in the repo. Removing it from here.

Re AI usage policy: PAC has the right to request/require changes be made to plugins in order for them to be published. Having AI usage be set accordingly as a prerequisite to submission should be acceptable.

I almost wonder, however, if we should have a more generic `disclaimers` list. This can include things like `ai_code_l4`, or `requires_network_access` as a special case of tag.

